PR #6, opened 2021-06-01
Hw dalai

Dalai-S:


src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            throw new DataProcessingException("Can't get a cinema hall by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("FROM CinemaHall", CinemaHall.class).getResultList();
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a cinema hall by id", e);
```
mankokolya, line 51: wrong message in exception

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,64 @@
...
+            transaction = session.beginTransaction();
+            session.save(movieSession);
+            transaction.commit();
+            return movieSession;
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException(
+                    "Can't insert movie session" + movieSession.getId(), e);
```
mankokolya, line 33: Why do you pass only id in the message? By the way, you will not have id if you get the exception. Pass movieSession here

mankokolya (changes requested):

mankokolya (approved):

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,90 @@
...
+
+@Entity
+@Table(name = "sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
```
RomanDubovskyi, line 20: default fetch type is eager here, let's use lazy with join fetch

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,90 @@
...
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
```
RomanDubovskyi, line 18: same

RomanDubovskyi (changes requested):

MaliukDaria (approved):

PR #33, opened 2021-06-06
completed task

mmelniichuk:


latkova (approved):

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.persistence.OneToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToOne
```
MaliukDaria, line 18: One movie could have only one movie session?

MaliukDaria (changes requested):

src/main/java/mate/academy/model/MovieSession.java
mmelniichuk: definitely not)

MaliukDaria (approved):

PR #75, opened 2021-09-11
Initial solution

VitaliyJvM:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> getMovieSessionQuery = session.createQuery("FROM MovieSession ms "
+                    + "LEFT JOIN FETCH ms.cinemaHall "
+                    + "WHERE ms.movie.id = :movieId AND ms.showTime BETWEEN :startDT AND :endDT",
+                    MovieSession.class);
+            getMovieSessionQuery.setParameter("movieId", movieId);
+            getMovieSessionQuery.setParameter("startDT",
+                    LocalDateTime.of(date, LocalTime.of(0,0,0)));
```
RomanDubovskyi, line 59: let's use constants here LocalTime.MAX / MIN for better readability

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,73 @@
...
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
```
RomanDubovskyi, line 19: default fetch type is eager here, since we use join fetch - let's set it to Lazy

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #79, opened 2021-09-28
hibernate-create-movie completed.

karaedd:


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
```
RomanDubovskyi, line 18: since we use join fetch in our hibernate queries it's better to use Lazy fetch type here. Since default one is eager

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,65 @@
...
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+        try (Session session = sessionFactory.openSession()) {
+            Query<MovieSession> getMovieSessionQuery = session.createQuery("FROM MovieSession ms "
+                    + "join fetch ms.movie m "
+                    + "WHERE m.id = :movieId AND ms.showTime"
+                    + " BETWEEN :startSession AND :endSession ", MovieSession.class);
+            getMovieSessionQuery.setParameter("movieId", movieId);
+            getMovieSessionQuery.setParameter("startSession", date.atStartOfDay());
+            getMovieSessionQuery.setParameter("endSession", date.atTime(23, 59, 59));
```
RomanDubovskyi, line 58: let's use here LocalTime.MAX constant for better readability

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #125, opened 2021-11-16
added solution

YaroslavYarmoshyk:


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
```
RomanDubovskyi, line 17: For relations: ...toOne the default fetchType is eager, it's considered to be a good practice to use lazy fetch type + join fetch

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #127, opened 2021-11-29
Movie session hw

7777vova7777:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a movie session by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("from MovieSession m "
+                            + "left join fetch m.movie "
```
RomanDubovskyi, line 52:
```suggestion
                            + "join fetch m.movie "
```
we can use inner join here, since there shouldn't be movie session without movie

RomanDubovskyi (approved):

PR #144, opened 2021-12-24
initial commit

SergiiSV:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,62 @@
...
+import mate.academy.exception.DataProcessingException;
+import mate.academy.lib.Dao;
+import mate.academy.model.MovieSession;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.query.Query;
+
+@Dao
+public class MovieSessionDaoImpl implements MovieSessionDao {
```
PineappleApplePinecone, line 16: Here you can create your sessionFactory like final field 'cause it is heavy object, and use everywhere below 
`private final SessionFactory factory = HibernateUtil.getSessionFactory();`

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+            throw new DataProcessingException("Can't get a CinemaHall by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<CinemaHall> getAllCinemaHallsQuery =
+                    session.createQuery("from CinemaHall", CinemaHall.class);
+            return getAllCinemaHallsQuery.getResultList();
```
PineappleApplePinecone, line 52: you can cut local variable here if you want to  `return session
                    .createQuery("FROM CinemaHall", CinemaHall.class)
                    .getResultList();`

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import org.hibernate.query.Query;
+
+@Dao
+public class CinemaHallDaoImpl implements CinemaHallDao {
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        Transaction transaction = null;
+        Session session = null;
+        try {
+            session = HibernateUtil.getSessionFactory().openSession();
```
PineappleApplePinecone, line 21: again you can add your SessionFactory object like constant field

PineappleApplePinecone (changes requested):

PineappleApplePinecone (approved):

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
SergiiSV: I agree, but I think this way is easier to read code

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
SergiiSV: I think this way is also ok

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,68 @@
...
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    @JoinColumn(name = "movie_id")
+    private Movie movie;
+    @OneToOne
```
issugy, line 22: why 1-1 ?

issugy (changes requested):

src/main/java/mate/academy/model/MovieSession.java
SergiiSV: at this moment one cinemaHall can have only one MovieSession. During the day it can have many.
As I understand when we talk about many smth to smth, means that at this moment it can have many. This is the same as a car has an engine at this moment and in 2 hours it has a new engine and will ManyToOne, but in this case, we always say OneToOne (car and engine, the car can have only one engine at this moment)
Am I not right?

issugy (approved):

PR #168, opened 2022-01-14
Wrote an assignment hibernate-create-movie-session-hw.

ASelivonchyk:
- configure hibernate.cfg.xml;
- created POJO CinemaHall, MovieSession;
- created CinemaHallDao, MovieSessionDao with its impl;
- implement getAll() method to MovieDao and its impl.


MaliukDaria (approved):
Good Job🚀

PR #170, opened 2022-01-29
solution hibernate-create-movie-session-hw

oleh-mazurkov:


src/main/java/mate/academy/dao/MovieSessionDao.java
```diff
@@ -0,0 +1,13 @@
+package mate.academy.dao;
+
+import java.time.LocalDate;
+import java.util.List;
+import mate.academy.model.MovieSession;
+
+public interface MovieSessionDao {
+    MovieSession add(MovieSession movieSession);
+
+    MovieSession get(Long id);
```
MaliukDaria, line 10:
```suggestion
    Optional<MovieSession> get(Long id);
```

lets use Optional here

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "movie_session")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @OneToOne
```
MaliukDaria, line 19: one movie could have only one movieSession?

MaliukDaria (changes requested):

src/main/java/mate/academy/model/MovieSession.java
oleh-mazurkov: realy it`s fun))

MaliukDaria: Yes, but not very profitable for the cinema 🙂

src/main/java/mate/academy/service/MovieSessionService.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.time.LocalDate;
+import java.util.List;
+import java.util.Optional;
+import mate.academy.model.MovieSession;
+
+public interface MovieSessionService {
+    MovieSession add(MovieSession movieSession);
+
+    Optional<MovieSession> get(Long id);
```
MaliukDaria, line 11:
```suggestion
    MovieSession get(Long id);
```

Dao should return Optional where we can get null, but services should return objects

MaliukDaria (changes requested):

MaliukDaria (changes requested):
test

src/main/java/mate/academy/service/MovieSessionService.java
oleh-mazurkov: )) I see that I need to review lesson about Optional...

MaliukDaria (approved):

PR #175, opened 2022-02-23
Implementation of HW Hibernate - create CinemaHall and MovieSession

araslav:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,65 @@
...
+            throw new DataProcessingException("Can't get a movie session by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> getAllMovieSessionsQuery =
+                    session.createQuery("from MovieSession ms "
+                    + "left join fetch ms.movie msm "
```
RomanDubovskyi, line 54: movieSession can't be without movie, so you can use a default, inner join here

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,64 @@
...
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
```
RomanDubovskyi, line 19: toOne relations have default eager fetchType, it's better to put here laze and use join fetch where needed

RomanDubovskyi (changes requested):

MaliukDaria (approved):

PR #196, opened 2022-04-10
Implemented all methods and classes

antonignatenko:


IgorZima90 (approved):

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,68 @@
...
+            return Optional.ofNullable(session.get(MovieSession.class, id));
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get movie session by id " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        LocalDateTime from = date.atStartOfDay();
+        LocalDateTime until = date.atTime(23,59,59);
```
KostyaNikiforov, line 53: You can use this method, to avoid magic numbers)
`LocalDateTime.of(date, LocalTime.MAX));`

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,68 @@
...
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> findAvailableSessionQuery =
+                    session.createQuery("FROM MovieSession ms JOIN FETCH ms.movie m"
+                            + " JOIN FETCH ms.cinemaHall ch "
+                            + "WHERE ms.movie.id =:id"
+                            + " AND ms.showTime >:startOfDay AND ms.showTime <:endOfDay ");
```
KostyaNikiforov, line 59:
```suggestion
                            + " AND ms.showTime >= :startOfDay AND ms.showTime <= :endOfDay ");
```
What if we have the movie session starts at 00:00, do we get it in the result?

KostyaNikiforov (changes requested):

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
antonignatenko: makes sense
already changed

KostyaNikiforov (approved):

PR #202, opened 2022-04-11
Task is done, working correctly, codestyle checked.

lugburz-hr:


DenTima (approved):
I think everything is great!

PR #247, opened 2022-06-21
created CinemaHall, MovieSession and their services and daos

viktoriia-haraniuk:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        LocalDateTime startOfDay = date.atTime(LocalTime.MIN);
+        LocalDateTime endOfDay = date.atTime(LocalTime.MAX);
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> query
+                    = session.createQuery("from MovieSession m where m.showTime "
+                    + "between :value1 and :value2", MovieSession.class);
+            query.setParameter("value1", startOfDay);
```
liliia-ponomarenko, line 58: how about movieId?

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
liliia-ponomarenko, line 22: what happens if we pass a non-existent value id?

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
liliia-ponomarenko, line 23: the same here

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
viktoriia-haraniuk: 🤦‍♀️

liliia-ponomarenko (approved):
great 🔥

PR #252, opened 2022-07-09
implement task

gdpbalt:
- add entities CinemaHall, MovieSession, Movie;
- add DAO interfaces and implementations for those entities;
- add services for work with those entities;
- add custom runtime exception DataProcessingException;
- add configuration for hibernate.

### DIFF
--- a/src/main/java/mate/academy/dao/MovieDao.java
+++ b/src/main/java/mate/academy/dao/MovieDao.java
@@ -5,7 +5,7 @@ import java.util.Optional;
 import mate.academy.model.Movie;
 
 public interface MovieDao {
-    Movie add(Movie movie);
+    Movie add(Movie entity);
 
     Optional<Movie> get(Long id);
 
--- a/src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
@@ -2,19 +2,19 @@ package mate.academy.dao.impl;
 
 import java.util.List;
 import java.util.Optional;
-import mate.academy.dao.MovieDao;
+import mate.academy.dao.CinemaHallDao;
 import mate.academy.exception.DataProcessingException;
 import mate.academy.lib.Dao;
-import mate.academy.model.Movie;
+import mate.academy.model.CinemaHall;
 import mate.academy.util.HibernateUtil;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
 import org.hibernate.query.Query;
 
 @Dao
-public class MovieDaoImpl implements MovieDao {
+public class CinemaHallDaoImpl implements CinemaHallDao {
     @Override
-    public Movie add(Movie entity) {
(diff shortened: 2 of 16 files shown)


src/main/java/mate/academy/dao/MovieDao.java
```diff
@@ -5,7 +5,7 @@
 import mate.academy.model.Movie;
 
 public interface MovieDao {
-    Movie add(Movie movie);
+    Movie add(Movie entity);
```
antonEEmus, line 8: Why did you rename this? It's better to differentiate entities.

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        LocalTime timeStartOfDay = LocalTime.of(0, 0, 0, 0);
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            String hql = "FROM MovieSession "
+                    + "WHERE movie.id = :id "
+                    + "AND showTime >= :begin AND showTime < :end";
```
antonEEmus, line 56: You can use BETWEEN here

antonEEmus (changes requested):

src/main/java/mate/academy/dao/MovieDao.java
gdpbalt: I saw that in previsious task
I can cut and past the code without changing

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
gdpbalt: Cause BETWEEN include second value
I don't want this behaviour

From documantation 
https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_between

antonEEmus (approved):

antonEEmus (approved):

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,57 @@
...
+            throw new DataProcessingException("Can't get a cinema hall by id " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            String hql = "FROM CinemaHall";
+            Query<CinemaHall> query = session.createQuery(hql, CinemaHall.class);
+            return query.getResultList();
```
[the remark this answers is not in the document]
KostyaNikiforov: Also `hql` is not the best option to name query strings. To make it more precise you can call it `hqlQuery` or just `query`

KostyaNikiforov (approved):

PR #261, opened 2022-07-19
dao and service were implemented, new models were created

mykola8426:


src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
+<hibernate-configuration>
+    <session-factory>
+        <property name="dialect">org.hibernate.dialect.MySQL5Dialect</property>
+        <property name="connection.url">jdbc:mysql://localhost:3306/ticket_app?serverTimeZone=UTC</property>
```
zim10093, line 4:
```suggestion
        <property name="connection.url">jdbc:mysql://localhost:3306/ticket_app?serverTimezone=UTC</property>

```
In your case it works, because your program can work without this parameter

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">root</property>
+        <property name="connection.password">12345789</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.model.Movie"/>
+        <mapping class="mate.academy.model.CinemaHall"/>
+        <mapping class="mate.academy.model.MovieSession"/>
+    </session-factory>
+</hibernate-configuration>
```
zim10093, line 15: I think better to add an empty line at the end for

src/main/java/mate/academy/Main.java
```diff
@@ -1,16 +1,55 @@
...
 import mate.academy.model.Movie;
+import mate.academy.model.MovieSession;
+import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
+import mate.academy.service.MovieSessionService;
 
 public class Main {
     public static void main(String[] args) {
-        MovieService movieService = null;
+        Injector injector = Injector.getInstance("mate.academy");
```
zim10093, line 16: Better move local field "injector" to class fields

zim10093 (approved):

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,53 @@
...
+            session.save(cinemaHall);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add cinema hall to db,"
+                    + " CinemaHall: " + cinemaHall, e);
+        }
+        return null;
```
liliia-ponomarenko, line 32: Why do you return null?

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,66 @@
...
+            session.save(movieSession);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add movie session to db,"
+                    + " MovieSession: " + movieSession, e);
+        }
+        return null;
```
liliia-ponomarenko, line 34: the same here

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,66 @@
...
+            throw new DataProcessingException("Can't get movie session by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long moveId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> query = session.createQuery(
+                    "FROM MovieSession m "
+                            + "LEFT JOIN FETCH m.movie "
```
liliia-ponomarenko, line 51: is it really necessary to join the movie table?

liliia-ponomarenko (changes requested):
let's improve  your solution ;)

liliia-ponomarenko (approved):
😊

PR #290, opened 2022-09-10
Hibernate-create-movie-session-hw-maxlut

maxlutovinov:


src/main/java/mate/academy/dao/GenericDao.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import mate.academy.exception.DataProcessingException;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.query.Query;
+
+public interface GenericDao<T> {
```
aaananas, line 12: good idea to use generic dao, but in tasks like this please don't use it - mentor can reject this solution because it was not the tak here
when you will move your project to github you can add GenericDao there

aaananas (approved):

src/main/java/mate/academy/dao/GenericDao.java
maxlutovinov: Ок, thank you for advice.

PR #300, opened 2022-09-24
HW hibernate-create-movie-session is done.

Valerii-Sh:


src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,57 @@
...
+            throw new DataProcessingException("Can't get a cinema hall by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<CinemaHall> getAllCinemaHallsQuery =
+                    session.createQuery("from CinemaHall", CinemaHall.class);
+            return getAllCinemaHallsQuery.getResultList();
```
aaananas, line 52:
```suggestion
            return session.createQuery("from CinemaHall", CinemaHall.class).getResultList();
```
you can also do it like this

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,66 @@
...
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
+    @JoinColumn(name = "cinema_hall_id")
+    private CinemaHall cinemaHall;
+    private LocalDateTime showTime;
```
aaananas, line 23:
```suggestion
    @Column(name = "show_time")
    private LocalDateTime showTime;
```

aaananas (approved):

PR #363, opened 2022-11-21
CinemaHall and MovieSession added

hpavlo:


nazariipertsak (approved):
Looks good!

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return Optional.ofNullable(session.get(CinemaHall.class, id));
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a cinema hall by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        String query = "select c from CinemaHall c";
```
liliia-ponomarenko, line 48:
```suggestion
        String query = "from CinemaHall";
```
It has to work 😉

liliia-ponomarenko (approved):
Awesome  😍

PR #365, opened 2022-11-24
Solution was added

Kostiantyn-Sacheiev:


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,73 @@
...
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToOne
+    private CinemaHall cinemaHall;
+    @OneToOne
+    private Movie movie;
```
aaananas, line 20: this would mean that one movie can be shown just once

aaananas (changes requested):

aaananas (approved):

PR #374, opened 2022-12-01
Implemented task

GrishaSeniv:


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
```
PavloPolovyi, line 20: I think it should be one to one

PavloPolovyi (changes requested):
Better change manytoone to onetoone in MovieSession relation to Movie, but if u not agree with me, write me, i'll approve

src/main/java/mate/academy/model/MovieSession.java
PavloPolovyi: Sorry, i made mistake, your solution is correct

PavloPolovyi (approved):

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,71 @@
...
+    @ManyToOne
+    @JoinColumn(name = "cinema_hall_id")
+    private CinemaHall cinemaHall;
```
hronograf, line 23: why do we have `JoinColumn` for `CinemaHall` and don't have for `Movie`?

src/main/java/mate/academy/dao/GenericDao.java
```diff
@@ -0,0 +1,12 @@
...
+import java.util.List;
+import java.util.Optional;
+
+public interface GenericDao<T> {
+    T add(T entity);
+
+    Optional<T> get(Long id);
+
+    List<T> getAll();
+}
```
hronograf, line 12: cool!

hronograf (changes requested):
Good job!

hronograf (approved):
Nice! 🍉

PR #390, opened 2022-12-06
Hibernate configured, All Services & Dao implemented

NSay0:


denys-domashevskyi (approved):
Seems like everything is good! Nice job)

pom.xml
```diff
@@ -19,6 +19,12 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
```
liliia-ponomarenko, line 24: 🔥

src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,19 @@
+package mate.academy.model;
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import lombok.Data;
+
+@Data
```
liliia-ponomarenko, line 10: better avoid using Data-annotation with an entity, read more about it https://www.jpa-buddy.com/blog/lombok-and-jpa-what-may-go-wrong/
but in general, Lombok is a nice tool.

liliia-ponomarenko (approved):
Good job! Hope the article will be useful for you.

PR #405, opened 2022-12-19
Jv hv

DooMLeoN:
sorry for the SELECT ch FROM and so on, but I'm very annoyed that IntelliJ IDEA writes that this is an incorrect formulation

### DIFF
--- a/src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
+++ b/src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
@@ -1,7 +1,7 @@
 package mate.academy.service.impl;
 
-import javax.persistence.EntityExistsException;
 import java.util.List;
+import javax.persistence.EntityExistsException;
 import mate.academy.dao.CinemaHallDao;
 import mate.academy.lib.Inject;
 import mate.academy.lib.Service;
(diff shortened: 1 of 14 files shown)


src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public CinemaHall add(CinemaHall movie) {
+        return cinemaHallDao.add(movie);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).orElseThrow(()
+                -> new EntityExistsException("Can't get cinema hall by id: " + id));
```
aaananas, line 24:
```suggestion
                -> new RuntimeException("Can't get cinema hall by id: " + id));
```

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).orElseThrow(()
+                -> new EntityExistsException("Can't get movie session by id: " + id));
```
aaananas, line 25:
```suggestion
                -> new RuntimeException("Can't get movie session by id: " + id));
```

aaananas (approved):

PR #408, opened 2023-01-07
Hw 30 hibernate create movie session hw

SichniyAndriy:


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,72 @@
...
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private LocalDateTime showTime;
+    @OneToOne
+    private CinemaHall cinemaHall;
+    @OneToOne
```
aaananas, line 20: eager is a bad practice

aaananas (changes requested):

aaananas (approved):

PR #454, opened 2023-03-01
feat: hibernate-create-movie-session #comment was added CinemaHall an…

sania517:
…d MovieSession models

### DIFF
--- a/src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
@@ -2,55 +2,53 @@ package mate.academy.dao.impl;
 
 import java.util.List;
 import java.util.Optional;
-import mate.academy.dao.MovieDao;
+import mate.academy.dao.CinemaHallDao;
 import mate.academy.exception.DataProcessingException;
 import mate.academy.lib.Dao;
-import mate.academy.model.Movie;
+import mate.academy.model.CinemaHall;
 import mate.academy.util.HibernateUtil;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
-import org.hibernate.query.Query;
 
 @Dao
-public class MovieDaoImpl implements MovieDao {
+public class CinemaHallDaoImpl implements CinemaHallDao {
     @Override
-    public Movie add(Movie movie) {
-        Transaction transaction = null;
+    public CinemaHall add(CinemaHall cinemaHall) {
         Session session = null;
+        Transaction transaction = null;
         try {
             session = HibernateUtil.getSessionFactory().openSession();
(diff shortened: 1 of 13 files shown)


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,82 @@
...
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return Optional.ofNullable(session.get(MovieSession.class, id));
+        } catch (Exception ex) {
+            throw new DataProcessingException("Can not get MovieSessiom by id. id: " + id, ex);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        if (movieId == null || date == null) {
```
olekskov, line 54: If you wish to add validations it should be done on before input reaches dao layer. Controller or service layer are more suitable for it

olekskov (changes requested):

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    public CinemaHall add(CinemaHall cinemaHall) {
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = HibernateUtil.getSessionFactory().openSession();
+            transaction = session.beginTransaction();
+            session.save(cinemaHall);
+            transaction.commit();
+        } catch (Exception e) {
+            transaction.rollback();
```
Ivan95kos, line 25: It is necessary to check whether the transaction is not null, as an error may occur when opening a session

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            LocalDateTime startOfDay = date.atStartOfDay();
+            LocalDateTime endOfDay = date.atTime(LocalTime.MAX);
+
+            Query<MovieSession> getMovieSessionListQuery = session.createQuery(
+                    "FROM MovieSession ms "
+                            + "WHERE ms.movie.id = :movieId "
+                            + "AND ms.showTime BETWEEN :startOfDay AND :endOfDay",
```
Ivan95kos, line 61: Better to create a variable for this query

Ivan95kos (changes requested):

uliana-andrushchuk (approved):

PR #463, opened 2023-03-07
main solution

ivan-vovnyanko:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        Optional<MovieSession> optionalMovieSession = movieSessionDao.get(id);
+        if (optionalMovieSession.isEmpty()) {
+            throw new NoSuchElementException("Couldn't find movie session by id " + id);
+        }
+        return optionalMovieSession.get();
+    }
```
okuzan, line 30:
```suggestion
        return optionalMovieSession.orElseThrow(() -> new NoSuchElementException("Couldn't find movie session by id " + id));
    }
```

src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +22,10 @@ public Optional<Movie> get(Long id) {
 
     @Override
     public List<Movie> getAll() {
-        return null;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("from Movie ", Movie.class).getResultList();
```
okuzan, line 26: better use upper case in queries

okuzan (changes requested):

Rommelua (approved):

PR #474, opened 2023-03-20
Implemented DAO, Service, model classes

andriy-linkov:


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,57 @@
...
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
+    private CinemaHall cinemaHall;
```
aaananas, line 20: eager is a bad practice - use Lazy and fetch relations

aaananas (changes requested):

Rommelua (approved):

PR #491, opened 2023-04-11
Initial commit

Dominskyi93:


src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -22,7 +24,7 @@ public Movie add(Movie movie) {
             session.save(movie);
             transaction.commit();
             return movie;
-        } catch (Exception e) {
+        } catch (RuntimeException e) {
```
karma-o, line 27: we should be able to catch and handle any exception and not just runtime, fix everywhere

karma-o (changes requested):

karma-o (approved):

PR #501, opened 2023-04-14
Movie session services implemented

IlkoGood:
LocalDateTime is still pain in the arse

### DIFF
--- a/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
@@ -1,0 +1,61 @@
+package mate.academy.dao.impl;
+
+import java.time.LocalDate;
+import java.util.List;
+import java.util.Optional;
+import mate.academy.dao.MovieSessionDao;
+import mate.academy.exception.DataProcessingException;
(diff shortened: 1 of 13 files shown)


V-Vasylchuk (approved):

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("from MovieSession ms where "
+                            + "ms.movie.id = " + movieId
+                            + " and year(ms.showTime) = " + date.getYear()
+                            + " and month(ms.showTime) = " + date.getMonthValue()
+                            + " and day(ms.showTime) = " + date.getDayOfMonth(),
```
karma-o, line 54: we should use `query.setParameter()` method instead to secure ourselves from the SQL injections, check other queries

karma-o (changes requested):

karma-o (approved):

PR #519, opened 2023-04-17
Hw

Deschna:


src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get the cinemaHall with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<CinemaHall> query = session.createQuery("FROM CinemaHall", CinemaHall.class);
+            return query.list();
```
oleksiistepaniak, line 38: You can inline this string of code instead of the existing one. For instance, you can write 'return session.createQuery("from CinemaHall", CinemaHall.class).getResultList(). It will be a little simplified, won't it?

src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +46,11 @@ public Optional<Movie> get(Long id) {
 
     @Override
     public List<Movie> getAll() {
-        return null;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Movie> query = session.createQuery("FROM Movie", Movie.class);
+            return query.list();
```
oleksiistepaniak, line 51: You can inline this string of code instead of the existing one. For instance, you can write 'return session.createQuery("from Movie", Movie.class).getResultList(). It will be a little simplified, won't it?

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.persistence.Id;
+import javax.persistence.OneToOne;
+
+@Entity
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
+    @OneToOne
+    private Movie movie;
```
oleksiistepaniak, line 16: In my point of view, it's better to use @ManyToOne here. Why? The explanation is next. A lot of movie sessions may have the same movie. Let's imagine that you are an owner of a cinema and your cinema has three movie sessions with the same movie and the same cinema hall on a particular day. For instance, it may be Home Alone 2. So, you may have three movie sessions with three identical movies and cinema halls. The difference between them will only be the time of beginning them. The first one will start at 9 o'clock, the second one will start at 10 o'clock, the third one will start at 11 o'clock and so on.

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,62 @@
...
+    @OneToOne
+    private CinemaHall cinemaHall;
```
oleksiistepaniak, line 18: Here is the same situation like above.

oleksiistepaniak (approved):
Overall, a good job. But I recommend finding out with my tips how to improve your solution a little bit.

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,41 @@
...
+import mate.academy.lib.Dao;
+import mate.academy.model.CinemaHall;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+
+@Dao
+public class CinemaHallDaoImpl implements CinemaHallDao {
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
```
karma-o, line 16: add transactions and use try-finally instead of try with resources in all non read-only methods

karma-o (changes requested):

karma-o (approved):

PR #528, opened 2023-04-18
hibernate.cfg.xml created

metelska-anastasiia:
created 2 model classes, dao layers and service layers for them


kozhukhovsky (approved):
Good job!

PR #534, opened 2023-04-22
created create CinemaHall and MovieSession (Hibernate)

vromanyuk:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+    }
+
+    @Override
+    public Optional<MovieSession> get(Long id) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> sessionQuery = session
+                    .createQuery("from MovieSession m "
+                            + "join fetch m.movie mm where m.id = :id ", MovieSession.class);
```
Ivan95kos, line 43: It is better to write all SQL commands in upper case, fix everywhere
```suggestion
                    .createQuery("FROM MovieSession m "
                            + "JOIN FETCH m.movie mm WHERE m.id = :id ", MovieSession.class);
```

Ivan95kos (changes requested):

Ivan95kos (approved):
Great job

PR #535, opened 2023-04-26
implemented new classes

andrey156:


src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,62 @@
...
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "cinema_halls")
+public class CinemaHall implements Cloneable {
```
Rommelua, line 11: Why did you decide to implement Cloneable? We do not need it.
For all models.

Rommelua (approved):

src/main/java/mate/academy/model/CinemaHall.java
andrey156: unfortunately, this was not a moderate decision, I was copying other models.

PR #549, opened 2023-05-28
solved task

AVykhovanok:


src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
olekskov, line 22: Please use .orElseThrow construction to produce proper exception with meaningful message in case entity is not present in DB.
Same for MovieSessionServiceImpl

olekskov (approved):
Good job overall, approved.
Please check comment below for further tasks

PR #566, opened 2023-06-23
task solution implemented

Tupadzaky:


MykytaKat (approved):
great job! keep pushing!

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,74 @@
...
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne(fetch = FetchType.LAZY)
+    private Movie movie;
+    @ManyToOne(fetch = FetchType.LAZY)
+    private CinemaHall cinemaHall;
+    private LocalDate showTime;
```
maxline, line 22: `showTime` field should be value of `LocalDateTime` type according the task description

maxline (changes requested):
Very good job! Left just one remark, fix it please.

src/main/java/mate/academy/Main.java
```diff
@@ -32,7 +32,7 @@ public static void main(String[] args) {
 
         MovieSessionService movieSessionService = (MovieSessionService)
                 injector.getInstance(MovieSessionService.class);
-        LocalDate localDate = LocalDate.of(2023, 6, 23);
+        LocalDateTime localDate = LocalDateTime.of(2023, 6, 23, 1, 1);
```
maxline, line 35: It is better to set more realistic time to test data:
`LocalDateTime.of(2023, 6, 23, 16, 30);`

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -53,7 +53,7 @@ public Optional<MovieSession> get(Long id) {
     }
 
     @Override
-    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDateTime date) {
```
maxline, line 56: This fix is redundant, according the task description date type should be:
`findAvailableSessions(Long movieId, LocalDate date) {`
But MovieSession  `showTime` should have `LocalDateTime` type.
Please fix and check that solution works.

maxline (changes requested):
We are close to finish. Added a couple more of minor remarks.

maxline (approved):
Well done! Approved

PR #567, opened 2023-06-23
File hibernate.cfg.xml was configured.

Ivan-Yuriv:
CinemaHall and MovieSession models were added to the project and their implementations on dao and service layers. All methods were tested in the Main.class.

### DIFF
--- a/src/main/java/mate/academy/model/MovieSession.java
+++ b/src/main/java/mate/academy/model/MovieSession.java
@@ -1,0 +1,73 @@
+package mate.academy.model;
+
+import java.time.LocalDateTime;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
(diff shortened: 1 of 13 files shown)


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,73 @@
...
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
+    private CinemaHall cinemaHall;
```
aaananas, line 20: eager is a bad practice

aaananas (changes requested):

aaananas (approved):

PR #568, opened 2023-06-26
hibernate-create-movie-session hw

kcasPilot:


src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
olekskov, line 22: Using Optional.get() without checking if the value is present may lead to NoSuchElementException. Consider using Optional.orElseThrow() with a proper exception instead.

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
olekskov, line 23: Using Optional.get() without checking if the value is present may lead to NoSuchElementException. Consider using Optional.orElseThrow() with a proper exception instead.

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        LocalDateTime startDay = date.atTime(0, 0, 0);
+        LocalDateTime endDay = date.atTime(23, 59, 59);
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> query = session.createQuery("from MovieSession ms "
+                    + "where showTime >= :startDay and showTime <= :endDay", MovieSession.class);
+            query.setParameter("startDay", startDay);
```
olekskov, line 56: Add 'query.setParameter("movieId", movieId);' after setting the 'endDay' parameter to include the movieId in the query.

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        LocalDateTime startDay = date.atTime(0, 0, 0);
+        LocalDateTime endDay = date.atTime(23, 59, 59);
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> query = session.createQuery("from MovieSession ms "
+                    + "where showTime >= :startDay and showTime <= :endDay", MovieSession.class);
```
olekskov, line 55: Add 'and ms.movie.id = :movieId' to the query to filter by movieId as required.

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            return Optional.ofNullable(session.get(CinemaHall.class, id));
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get CinemaHall with id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("FROM CinemaHall", CinemaHall.class).list();
```
olekskov, line 49: Replace 'list()' with 'getResultList()' to avoid using deprecated methods

olekskov (changes requested):

kozhukhovsky (approved):

PR #575, opened 2023-07-03
requirements were met

vgladkyi:


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,72 @@
...
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
```
overpathz, line 19: Eager fetching is a bad practice

overpathz (approved):
Nice, approved. But there is a thing not to do this in the future.

PR #582, opened 2023-07-06
Created MovieSession, CinemaHall models with their DAO and service im…

viprko:
…plementations.

### DIFF
--- a/src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
@@ -2,32 +2,32 @@ package mate.academy.dao.impl;
 
 import java.util.List;
 import java.util.Optional;
-import mate.academy.dao.MovieDao;
+import mate.academy.dao.CinemaHallDao;
 import mate.academy.exception.DataProcessingException;
 import mate.academy.lib.Dao;
-import mate.academy.model.Movie;
+import mate.academy.model.CinemaHall;
 import mate.academy.util.HibernateUtil;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
 import org.hibernate.query.Query;
 
 @Dao
-public class MovieDaoImpl implements MovieDao {
+public class CinemaHallDaoImpl implements CinemaHallDao {
     @Override
-    public Movie add(Movie movie) {
+    public CinemaHall add(CinemaHall cinemaHall) {
         Transaction transaction = null;
         Session session = null;
         try {
             session = HibernateUtil.getSessionFactory().openSession();
             transaction = session.beginTransaction();
-            session.save(movie);
+            session.save(cinemaHall);
             transaction.commit();
-            return movie;
+            return cinemaHall;
         } catch (Exception e) {
             if (transaction != null) {
                 transaction.rollback();
             }
-            throw new DataProcessingException("Can't insert movie " + movie, e);
+            throw new DataProcessingException("Can't insert cinema hall " + cinemaHall, e);
         } finally {
             if (session != null) {
                 session.close();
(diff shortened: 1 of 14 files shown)


Pkrasota (approved):
Really cool work, nothing to add 👍

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,57 @@
...
+            throw new DataProcessingException("Can't get a cinema hall by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<CinemaHall> getAllCinemaHallsQuery =
+                    session.createQuery("from CinemaHall ", CinemaHall.class);
+            return getAllCinemaHallsQuery.getResultList();
```
sarakhmen, line 52: I'd recommend inlining it. You may fix the same issue in all places

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,76 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    @JoinColumn(name = "movie_id")
+    private Movie movie;
+    @ManyToOne
+    @JoinColumn(name = "cinema_hall_id")
+    /*@JoinTable(name = "movie_sessions_cinema_halls",
+    joinColumns = @JoinColumn(name = "movie_session_id"),
+    inverseJoinColumns = @JoinColumn(name = "cinema_halls_id"))*/
```
sarakhmen, line 25: let's remove redundant comments

src/main/java/mate/academy/lib/Injector.java
```diff
@@ -33,6 +33,16 @@ public static Injector getInstance(String mainPackageName) {
...
     }
 
+    /**
+     * Scans all classes accessible from the context class loader which
+     * belong to the given package and subpackages.
+     *
+     * @param packageName The base package
+     * @return The classes
+     * @throws ClassNotFoundException if the class cannot be located
+     * @throws IOException            if I/O errors occur
```
sarakhmen, line 43: please revert all the changes applied to Injector class

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,76 @@
...
+        this.showTime = showTime;
+    }
+
+    @Override
+    public String toString() {
+        return "MovieSession{" + "id=" + id + ", movie=" + movie + ", cinemaHall=" + cinemaHall
+                + ", showTime=" + showTime + '}';
+    }
+}
+
```
sarakhmen, line 76:
```suggestion
```

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id)
+                .orElseThrow(() -> new NoSuchElementException("Can't get cinema hall"));
```
sarakhmen, line 24: let's also specify the ID in your exception message here. Fix the same issue in all places

sarakhmen (changes requested):
Let's improve your solution :)

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
viprko: But isn't it a violation of security or a violation of encapsulation? I mean, aren't we throwing exceptions on the service layer to the end user? So he shouldn't know about redundant information for him, like the id of something?

maxline: 1. It depends of an specific application, for some entities id can be hidden for some not. E.g. you can often see example of the REST request get user by id, where the `id` is a part of the url parameter.
2. Throwing an exception here doesn't mean that exception message will be shown to the end user, message will be put to the application log, and should have contain all details for possible investigation the problem.

maxline (approved):
Well done! Approved

PR #585, opened 2023-07-07
Homework: Hibernate Create Movie Session

ViktorBykh:


src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,76 @@
...
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToOne
+    private Movie movie;
+    @OneToOne
+    @JoinColumn(name = "cinema_hall_id")
+    private CinemaHall cinemaHall;
```
chrisnyrka, line 23: I think the  Many-to-One relation is better here (and note it is LAZY by default)

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
+    }
```
chrisnyrka, line 24: Here you need to use orElseThrow(...) 
Check in all cases

chrisnyrka (approved):
Good job) A few recommendations

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,84 @@
...
+            Query<MovieSession> getAllMovieSessionsQuery =
+                    session.createQuery("FROM MovieSession ms "
+                            + "LEFT JOIN FETCH ms.cinemaHall "
+                            + "LEFT JOIN FETCH ms.movie "
+                            + "WHERE ms.movie.id = :movieId "
+                            + "AND YEAR(ms.showTime) = :year "
+                            + "AND MONTH(ms.showTime) = :month "
+                            + "AND DAY(ms.showTime) = :day "
+                            + "AND ms.showTime "
+                            + "BETWEEN :startOfDay AND :endOfDay",
```
sarakhmen, line 68: I think this condition is redundant here :)

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #609, opened 2023-07-11
movie session / home-work

andrii-subash:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,70 @@
...
+            return Optional.ofNullable(session.get(MovieSession.class, id));
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a movie session by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        LocalDateTime beginningOfDay = date.atTime(0, 0, 0);
+        LocalDateTime endOfDay = date.atTime(23, 59, 59);
```
Ivan95kos, line 53: It is better to use next examples:
```
LocalDateTime fromTime = date.atStartOfDay();
LocalDateTime toTime = date.atTime(LocalTime.MAX);
```

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> movieSessionQuery =
+                    session.createQuery("from MovieSession ms "
```
Ivan95kos, line 57: It is better to write all SQL commands in upper case and fix them everywhere
```suggestion
                    session.createQuery("FROM MovieSession ms "
```

Ivan95kos (changes requested):
Good job overall!

PR #612, opened 2023-07-14
Hw branch

yaaariiik:


src/main/java/mate/academy/Main.java
```diff
@@ -1,16 +1,76 @@
...
+
+        MovieService movieService = (MovieService) injector.getInstance(MovieService.class);
 
         Movie fastAndFurious = new Movie("Fast and Furious");
         fastAndFurious.setDescription("An action film about street racing, heists, and spies.");
         movieService.add(fastAndFurious);
         System.out.println(movieService.get(fastAndFurious.getId()));
+
+        Movie anotherFilm = new Movie("SomeFilm");
+        fastAndFurious.setDescription("Something happens in this film.");
```
olekskov, line 25: You are setting the description of the wrong movie. It should be 'anotherFilm.setDescription', not 'fastAndFurious.setDescription'.

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> getAllAvailableSessionsQuery = session.createQuery(
+                    "FROM MovieSession ms "
+                    + "JOIN FETCH ms.movie m "
+                    + "JOIN FETCH ms.cinemaHall "
+                    + "WHERE EXTRACT(YEAR FROM ms.showTime) = :year "
```
olekskov, line 56: The handling of date fields can be simplified by comparing directly with a date object, rather than extracting and comparing year, month, and day separately.

src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +47,11 @@ public Optional<Movie> get(Long id) {
```
olekskov: After throwing DataProcessingException, it's necessary to roll back transaction. Add transaction.rollback() before throwing exception.

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,75 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    @JoinTable(name = "sessions_movies",
+            joinColumns = @JoinColumn(name = "movie_session_id"),
+            inverseJoinColumns = @JoinColumn(name = "movie_id"))
+    private Movie movie;
+    @ManyToOne
+    @JoinTable(name = "sessions_halls",
```
olekskov, line 25: For many-to-one relation it's surely possible to use jointable but having just column with foreign key is more simple solution and fully suitable here

src/main/java/mate/academy/dao/CinemaHallDao.java
```diff
@@ -0,0 +1,8 @@
+package mate.academy.dao;
+
+import java.util.List;
+import mate.academy.model.CinemaHall;
+
+public interface CinemaHallDao extends GenericDao<CinemaHall> {
+    List<CinemaHall> getAll();
```
olekskov, line 7: If you introduced GenericDao why getAll() method is not included there but separately hardcoded in each dao?

olekskov (changes requested):

src/main/java/mate/academy/dao/CinemaHallDao.java
yaaariiik: I have three dao interfaces that extends GenericDao, and according to requirements only two of them have getAll method. If i'll add getAll() method to GenericDao, then I need to implement it in MovieSessionDao, but README.md tells me that it should have only add, get and findAvailableSessions methods. In a list with common mistakes I have recomendation "Do not create redundant functionality, like methods that are not required by the task." So, that's why I did as I did :) The same with Service layer.

kozhukhovsky (approved):

PR #618, opened 2023-07-28
task hibernate-create-movie-session is completed

DenisKalyakin:


src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hibernate.query.Query;
+
+@Dao
+public class CinemaHallDaoImpl implements CinemaHallDao {
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        Transaction transaction = null;
+        Session session = null;
+        try {
+            session = HibernateUtil.getSessionFactory().openSession();
```
sokimaaa, line 21: you can use try-with-resources for autoclose the session

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,67 @@
...
+import org.hibernate.query.Query;
+
+@Dao
+public class MovieSessionDaoImpl implements MovieSessionDao {
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        Transaction transaction = null;
+        Session session = null;
+        try {
+            session = HibernateUtil.getSessionFactory().openSession();
```
sokimaaa, line 23: you can use here `try-with-resources` for autoclosing the session

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
sokimaaa, line 22: directly calling the `Optional#get` without a previous empty check is bad practice (and may produce Exception in case of empty).

Pls, do an empty check, or use another option such as `Optional#orElse`, `Optional#orElseThrow`.

sokimaaa (changes requested):

sokimaaa (approved):

PR #620, opened 2023-07-29
all done

oleksandrCod:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
olekskov, line 23: Please use .orElseThrow() method with meaningful exception instead of raw usage or .get() to produce more understandable result when entity wasn't found.
Please recheck all occurences

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,87 @@
...
+        Session session = null;
+        try {
+            session = factory.openSession();
+            Query<MovieSession> allMovieSession = session.createQuery("from MovieSession ms "
+                    + " left join fetch ms.cinemaHall "
+                    + " left join fetch ms.movie"
+                    + " where ms.id = :id", MovieSession.class);
+            allMovieSession.setParameter("id", id);
+            return allMovieSession.uniqueResultOptional();
+        } catch (Exception e) {
```
olekskov, line 55: Please use try-with-resources for all get(), getAll() operations

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,87 @@
...
+        return movieSession;
+    }
+
+    @Override
+    public Optional<MovieSession> get(Long id) {
+        SessionFactory factory = HibernateUtil.getSessionFactory();
+        Session session = null;
+        try {
+            session = factory.openSession();
+            Query<MovieSession> allMovieSession = session.createQuery("from MovieSession ms "
```
olekskov, line 49: Why all? We're preparing query for particular single entity
```suggestion
            Query<MovieSession> movieSessionById = session.createQuery("from MovieSession ms "

```

olekskov (changes requested):

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
oleksandrCod: Hi, why i need to use tryResources, i mean it use more perfomance or it the same as finally?

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,73 @@
...
+                session.close();
+            }
+        }
+        return movieSession;
+    }
+
+    @Override
+    public Optional<MovieSession> get(Long id) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> movieSessionById = session.createQuery("from MovieSession ms "
```
Ivan95kos, line 46: It is better to write all SQL commands in upper case
```suggestion
            Query<MovieSession> movieSessionById = session.createQuery("FROM MovieSession ms "

```

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
Ivan95kos: Both reasons: https://www.baeldung.com/java-try-with-resources

Ivan95kos (approved):

PR #633, opened 2023-08-14
Hibernate create CinemaHall and MovieSession

knyazkyylubomir:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> getMovieSessionQuery = session.createQuery("from MovieSession mv "
+                    + "where mv.movie.id = :movieId "
+                    + "and mv.showTime between :start and :end", MovieSession.class);
+            getMovieSessionQuery.setParameter("movieId", movieId);
+            getMovieSessionQuery.setParameter("start", date.atStartOfDay());
+            getMovieSessionQuery.setParameter(
+                    "end", date.atTime(TWENTY_THREE, FIFTY_NINE, FIFTY_NINE));
```
overpathz, line 60: It's better to move this values as variables in the top of the method. Smth like that:
```
LocalDateTime startOfDay = date.atStartOfDay();
LocalDateTime endOfDay = date.atTime(LocalTime.MAX);
```

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
overpathz, line 22: What if we do not find any object?

```
return cinemaHallDao.get(id).orElseThrow(
                () -> new NoSuchElementException("Can not get cinema hall with id:" + id));
    }
```

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
overpathz, line 23: Please, add orElseThrow

overpathz (changes requested):
In general ok, need some fixes

olekskov (approved):
Good job

PR #636, opened 2023-08-16
implemented dao and service sides for Movies, MovieSessions and Cinem…

holsoni:
…aHalls

### DIFF
--- a/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
@@ -1,0 +1,75 @@
+package mate.academy.dao.impl;
+
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.academy.dao.MovieSessionDao;
+import mate.academy.exception.DataProcessingException;
+import mate.academy.lib.Dao;
+import mate.academy.model.MovieSession;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.query.Query;
+
+@Dao
+public class MovieSessionDaoImpl implements MovieSessionDao {
+    private static final String MOVIE_ID = "movieId";
+    private static final String START_OF_DATE = "startOfDate";
+    private static final String END_OF_DATE = "endOfDate";
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        Transaction transaction = null;
+        Session session = null;
+        try {
+            session = HibernateUtil.getSessionFactory().openSession();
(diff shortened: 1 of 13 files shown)


Denis-Balako (changes requested):
Awesome solution! One minor issue needs want to be fixed 😺

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a movie session by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        String query = String.format("from MovieSession m where m.movie.id = :%s"
+                        + " and m.showTime >= :%s and m.showTime <= :%s",
+                "MOVIE_ID", "START_OF_DATE", "END_OF_DATE");
```
nacenik, line 57: now you get this string query: `from MovieSession m where m.movie.id = :MOVIE_ID and m.showTime >= :START_OF_DATE and` m.showTime <= :END_OF_DATE`
But I think it should be another 
```suggestion
        String query = String.format("from MovieSession m where m.movie.id = :%s"
                        + " and m.showTime >= :%s and m.showTime <= :%s",
                MOVIE_ID, START_OF_DATE, END_OF_DATE);
```

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,72 @@
...
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
```
nacenik, line 19: is it a good idea to use fetch type eager?

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> getAll = session.createQuery(query, MovieSession.class);
+            getAll.setParameter(MOVIE_ID, movieId);
+            getAll.setParameter(START_OF_DATE, date.atStartOfDay());
+            getAll.setParameter(END_OF_DATE, date.atTime(LocalTime.MAX));
+            return getAll.getResultList();
+        } catch (Exception e) {
+            throw new DataProcessingException(
+                    "Can't get movie session on required date: " + date, e);
```
nacenik, line 66: movie session for which movie id?

nacenik (changes requested):
Good solution, let's improve it a little bit :)

PR #644, opened 2023-08-17
created models of cinema hall and MovieSession , created dao levels a…

0-zagar-0:
…nd service level for these classes , tested all levels in main method Ð

### DIFF
--- a/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
@@ -1,0 +1,83 @@
+package mate.academy.dao.impl;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.academy.dao.MovieSessionDao;
+import mate.academy.exception.DataProcessingException;
+import mate.academy.lib.Dao;
+import mate.academy.model.MovieSession;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.query.Query;
+
+@Dao
+public class MovieSessionDaoImpl implements MovieSessionDao {
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        Session session = null;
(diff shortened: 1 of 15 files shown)


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
okuzan, line 23: orElseThrow nosuchelement

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
okuzan, line 22: same

okuzan (changes requested):

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,81 @@
...
+                session.close();
+            }
+        }
+        return movieSession;
+    }
+
+    @Override
+    public Optional<MovieSession> get(Long id) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> getMovieSessionQuery = session.createQuery("from MovieSession ms "
```
Ivan95kos, line 45: It is better to write all SQL commands in upper case and fix them everywhere
```suggestion
            Query<MovieSession> getMovieSessionQuery = session.createQuery("FROM MovieSession ms "

```

Ivan95kos (approved):

PR #653, opened 2023-08-22
cinemaHall and movieSession entities and their relations created, dao…

TymurProkhorov:
… and service methods implemented

### DIFF
--- a/src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
+++ b/src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
@@ -1,0 +1,29 @@
+package mate.academy.service.impl;
+
+import java.util.List;
+import mate.academy.dao.CinemaHallDao;
+import mate.academy.lib.Inject;
+import mate.academy.lib.Service;
+import mate.academy.model.CinemaHall;
+import mate.academy.service.CinemaHallService;
+
+@Service
+public class CinemaHallServiceImpl implements CinemaHallService {
+    @Inject
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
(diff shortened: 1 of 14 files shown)


src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
+    }
```
olekskov, line 23:
```suggestion
    @Override
    public CinemaHall get(Long id) {
        return cinemaHallDao.get(id).orElseThrow(new EntityNotFoundException("CinemaHall with id " + id + " doesn't exist"));
    }

```

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
+    }
```
olekskov, line 24: same here

Do not use just get() method of Optional. It makes no sense, we may obtain NPE same as when we return null

olekskov (approved):
Good job overall. Approved.
Please pay attention to comment below to act correctly further

PR #661, opened 2023-08-27
hibernate-create-movie-session

s-hrynus:
…oImpl and MovieSessionDaoImpl. Fixed LazyInitializationException in get() and findAvailableSessions() methods.

### DIFF
--- a/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
@@ -1,0 +1,74 @@
+package mate.academy.dao.impl;
+
+import java.time.LocalDate;
+import java.util.List;
+import java.util.Optional;
+import mate.academy.dao.MovieSessionDao;
+import mate.academy.exception.DataProcessingException;
+import mate.academy.lib.Dao;
+import mate.academy.model.MovieSession;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.query.Query;
+
+@Dao
+public class MovieSessionDaoImpl implements MovieSessionDao {
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = HibernateUtil.getSessionFactory().openSession();
+            transaction = session.beginTransaction();
+            session.persist(movieSession);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add movie session " + movieSession
+                    + " to DB", e);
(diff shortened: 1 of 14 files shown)


src/main/java/mate/academy/service/impl/MovieServiceImpl.java
```diff
@@ -19,11 +20,12 @@ public Movie add(Movie movie) {
 
     @Override
     public Movie get(Long id) {
-        return movieDao.get(id).get();
+        return movieDao.get(id).orElseThrow(
+                () -> new DataProcessingException("Can't get movie with id " + id));
```
okuzan, line 24: use nosuchelement or entitynotfound

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).orElseThrow(
+                () -> new DataProcessingException("Can't get cinema hall with id " + id));
```
okuzan, line 24: same

src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,53 @@
...
+public class CinemaHall {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private int capacity;
+    private String description;
+
+    public CinemaHall() {
+    }
+
```
okuzan, line 20: default constructor is created by default

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,74 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a movie session by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> getFilteredMovieSessionQuery = session
+                    .createQuery("from MovieSession ms "
```
okuzan, line 58:
```suggestion
                    .createQuery("FROM MovieSession ms "
```

okuzan (changes requested):

src/main/java/mate/academy/model/CinemaHall.java
s-hrynus: > default constructor is created by default

Yes, but if I delete it in all models, I will have an InstantiationException.

okuzan (approved):
Good job!

PR #668, opened 2023-09-05
Solution of this task

IvanovYurii:


src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
kozhukhovsky, line 22:
```suggestion
        return cinemaHallDao.get(id).orElseThrow(throw exception....);

```

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
kozhukhovsky, line 23: same

kozhukhovsky (approved):

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
IvanovYurii: > same

In the movie method, which was downloaded from GitHub, it is written like this. I decided that it is not necessary to change it and write as it is in it

PR #675, opened 2023-09-11
added hw solution hb-create-movie

vitaliiplys:


src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +46,11 @@ public Optional<Movie> get(Long id) {
 
     @Override
     public List<Movie> getAll() {
-        return null;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Movie> getAllMovies = session.createQuery("from Movie ");
```
Ivan95kos, line 50: It is better to write all SQL commands in upper case
```suggestion
            Query<Movie> getAllMovies = session.createQuery("FROM Movie ");
```

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,65 @@
...
+            return Optional.ofNullable(session.get(MovieSession.class,id));
+        } catch (Exception e) {
+            throw new DataProcessingException("Can`t get a movieSession by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> query = session.createQuery("from MovieSession ms "
```
Ivan95kos, line 52: the same

Ivan95kos (approved):

PR #677, opened 2023-09-12
completed task

BakhmetIvan:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
olekskov, line 23: Please do not use just .get() on Optionals. With such approach you will get same NPE as if you returned null and it's making Optional meaningless
Often it will be ok to use .orElseThrow() method with proper Exception type and with some meaningful message

olekskov (approved):
Good job overall, approved.
Please pay attention to comment below for further

olekskov (approved):

PR #680, opened 2023-09-16
first commit

chertterner:


src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,57 @@
...
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add cinemaHall to DB", e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+        return cinemaHall;
+    }
```
sokimaaa, line 36:
```suggestion
    public CinemaHall add(CinemaHall cinemaHall) {
        Session session = null;
        Transaction transaction = null;
        try {
            session = HibernateUtil.getSessionFactory().openSession();
            transaction = session.beginTransaction();
            session.persist(cinemaHall);
            transaction.commit();
            return cinemaHall;
        } catch (Exception e) {
            if (transaction != null) {
                transaction.rollback();
            }
            throw new DataProcessingException("Can't add cinemaHall to DB", e);
        } finally {
            if (session != null) {
                session.close();
            }
        }
    }
```

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,70 @@
...
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add movieSession to DB", e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+        return movieSession;
+    }
```
sokimaaa, line 37:
```suggestion
    public MovieSession add(MovieSession movieSession) {
        Session session = null;
        Transaction transaction = null;
        try {
            session = HibernateUtil.getSessionFactory().openSession();
            transaction = session.beginTransaction();
            session.persist(movieSession);
            transaction.commit();
            return movieSession;
        } catch (Exception e) {
            if (transaction != null) {
                transaction.rollback();
            }
            throw new DataProcessingException("Can't add movieSession to DB", e);
        } finally {
            if (session != null) {
                session.close();
            }
        }
    }
```

sokimaaa (changes requested):

src/main/java/mate/academy/model/Movie.java
```diff
@@ -4,10 +4,8 @@
 import jakarta.persistence.GeneratedValue;
 import jakarta.persistence.GenerationType;
 import jakarta.persistence.Id;
-import jakarta.persistence.Table;
 
-@Entity
-@Table(name = "movies")
+@Entity(name = "movies")
```
aaananas, line 8: entity name and table name are not the same
you should read about it, or try creating entities, with defined entity name, and other entity with defined table name

aaananas (changes requested):

sokimaaa (approved):

PR #684, opened 2023-09-25
added CinemaHall and MovieSession entities, their services and daos

RedArtelerist:


src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,66 @@
...
+import jakarta.persistence.Table;
+import java.util.List;
+
+@Entity
+@Table(name = "cinema_halls")
+public class CinemaHall {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
```
Andriy-dz, line 18: empty lines

src/main/java/mate/academy/dao/impl/AbstractDao.java
```diff
@@ -0,0 +1,31 @@
+package mate.academy.dao.impl;
+
+import java.util.Optional;
+import mate.academy.exception.DataProcessingException;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+
+public abstract class AbstractDao {
+    protected <T> T add(T entity) {
+        HibernateUtil.getSessionFactory().inTransaction(session -> {
```
Andriy-dz, line 10: dont use this method. Same in other casses

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,79 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @ManyToOne
+    @JoinColumn(name = "movie_id")
+    private Movie movie;
+
+    @ManyToOne
+    @JoinColumn(name = "cinema_hall_id")
+    private CinemaHall cinemaHall;
```
Andriy-dz, line 25: do you really need two-way connections in this case

Andriy-dz (changes requested):
Good job! Lets improve your solution)

src/main/java/mate/academy/model/MovieSession.java
RedArtelerist: I added bidirectional relational for cascade remove. I through that it was logical that MovieSession can't exist without CinemaHall and Movie like in real world. I didn’t see in the task requirements that I couldn’t do that, so why not :)

src/main/java/mate/academy/dao/impl/AbstractDao.java
RedArtelerist: Why not?) It does the same thing under the hood as I would write it usually, but ok, I replaced it.

Andriy-dz (approved):

PR #685, opened 2023-09-26
Implemented Model, Dao layer, Service layer, Main method

ystankevych:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,48 @@
...
+
+@Dao
+public class MovieSessionDaoImpl implements MovieSessionDao {
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        try {
+            return HibernateUtil.getFactory().fromTransaction(session -> {
+                session.persist(movieSession);
+                return movieSession;
+            });
```
RedArtelerist, line 20: Maybe inTransaction we need to use or it works in similar way? I confused because in another DAO you use inTransaction, so if it’s the same, then it’s better to write in the same style

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,72 @@
...
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    private Movie movie;
+    @ManyToOne
+    private CinemaHall cinemaHall;
```
RedArtelerist, line 20: It's good to add JoinColumn everywhere in ManyToOne relations. 

Also you can think about cascades (Perhaps if we delete a film, then it would be good to delete the sessions, which would not exist without them.) It's not obligatory, more to think from business logic's side

RedArtelerist (changes requested):
Great job, I left some minor comments to improve this solution.

src/main/java/mate/academy/model/MovieSession.java
ystankevych: Yes, I missed this moment (about joincolumns)
since we have a project diagram and bidirectional relations is not indicated on it I leave it as it is (but yes, if the diagram shows bidirectional relations, a cascade is possible).

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
ystankevych: Almost the same. But maybe you are right. It would be better to write in the same style =)

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try {
+            return HibernateUtil.getFactory()
+                    .fromSession(session -> session.createQuery("FROM MovieSession ms "
+            + "WHERE ms.movie.id = :movieId AND DATE(showTime) = :date", MovieSession.class)
+                    .setParameter("movieId", movieId).setParameter("date", date).getResultList());
```
NazarSmith228, line 41: would it make sense to extract this complex lambda to a separate method? 

`.fromSession(session -> findAvailableSessions(session, movideId, date)`

```
private List<MovieSession> findAvailableSessions(Session session, Long movideId, LocalDate date) {
// content from your lambda
}

File: src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java [hunk 4]
```diff
@@ -0,0 +1,47 @@
...
+                    Optional.ofNullable(session.get(MovieSession.class, id)));
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get movieSession by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try {
+            return HibernateUtil.getFactory()
```
NazarSmith228, line 38: one question regarding this query - have you checked its performance? will Hibernate fire a single SQL query to database with joins for movies and cinema halls?

by default `@ManyToOne` is eager, which means that for each session movie and hall will be loaded

have you tried this query with `join fetch` instead of eager fetch type? what will be the result?

always guestion yourself when working with a bit more complex relations :)

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,47 @@
...
+import mate.academy.model.MovieSession;
+import mate.academy.util.HibernateUtil;
+
+@Dao
+public class MovieSessionDaoImpl implements MovieSessionDao {
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        try {
+            HibernateUtil.getFactory().inTransaction(session ->
+                    session.persist(movieSession));
```
NazarSmith228, line 18: minor 

you could either inline lamda or move `inTransaction` to a new line

it would be beneficial for readability

**Think of all places, where you could improve readability**

NazarSmith228 (changes requested):
Great job! Just some minor remarks to fix

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
ystankevych: You mean explicitly set FetchType.Lazy for Movie and CinemaHall? I added join fetch to the query but if you mean to set lazy fetch type for movie and cinemaHall I  will make some changes.

Elena-Bruyako (approved):

PR #699, opened 2023-10-08
Hw solution

dmytro-voronin:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import org.hibernate.Transaction;
+
+@Dao
+public class MovieSessionDaoImpl implements MovieSessionDao {
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            String availableSessionsQueryFormat = """
+                    from MovieSession ms
```
Ivan95kos, line 21: It is better to write all SQL commands in upper case
```suggestion
                    FROM MovieSession ms
```

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public Optional<CinemaHall> get(Long id) {
+        return cinemaHallDao.get(id);
+    }
```
Ivan95kos, line 24:
```suggestion
    public CinemaHall get(Long id) {
        return cinemaHallDao.get(id).orElseThrow(() -> 
                    new EntityNotFoundException("Can't get CinemaHall by id: " + id));
    }
```

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    @Inject
+    private MovieSessionDao sessionDao;
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        return sessionDao.findAvailableSessions(movieId, date);
+    }
+
+    @Override
+    public Optional<MovieSession> get(Long id) {
```
Ivan95kos, line 23: the same

Ivan95kos (changes requested):
left a few comments

Ivan95kos (changes requested):
left a few comments

kozhukhovsky (approved):

PR #708, opened 2023-10-26
solution

kxnxc:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id);
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        List<MovieSession> movieSessionList = movieSessionDao.getAll();
+        return movieSessionList.stream()
+                .filter(ms -> ms.getId().equals(movieId)
+                        && ms.getShowTime().getDayOfYear() == date.getDayOfYear())
```
kerrrusha, line 32: would be nice to extract that to the Predicate<MovieSession> or to the method

kerrrusha (approved):

PR #709, opened 2023-10-27
Initial commit

gordgelin:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).orElseThrow(()
+                -> new DataProcessingException("Can't get a movie session by id: " + id));
```
olekskov, line 25: DataProcessingException indicates problems with DB (lost connection, problems with syntax, etc.)
Here operation worked find and we just don't have such session in DB so here should be EntityNotFoundExceprion or NoSuchElementException etc.

olekskov (approved):
Good job, approved.
Please pay attention to comment for further tasks

PR #712, opened 2023-10-31
added CinemaHall and implementations , added MovieSession and impleme…

vkucher49:
…ntations, added relations

### DIFF
--- a/src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
+++ b/src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
@@ -1,0 +1,35 @@
+package mate.academy.service.impl;
+
+import java.util.List;
+import java.util.Optional;
+import mate.academy.dao.CinemaHallDao;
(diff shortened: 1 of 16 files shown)


src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        Optional<CinemaHall> cinemaHallOptional = cinemaHallDao.get(id);
+        if (cinemaHallOptional.isEmpty()) {
+            throw new DataProcessingException("CinemaHall with ID: " + id + " not found");
+        }
+        return cinemaHallOptional.get();
```
kerrrusha, line 28: your variant is ok, also we can do more compact with inline `.orElseThrow(() -> new RuntimeException("..."));`

kerrrusha (approved):

PR #716, opened 2023-11-08
done

artificialLeo:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).orElse(null);
+    }
```
Sanderfuck, line 24: orElse(null); not good idea return null use (orElseThrow())

Sanderfuck (changes requested):

Rommelua (approved):

PR #717, opened 2023-11-10
solved hibernate create movie session

Krasnov-Maksim:


src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,65 @@
...
+            final Query<CinemaHall> query = session.createQuery(sqlText, CinemaHall.class);
+            result = query.list();
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't get a list of cinema halls.", e);
+        }
+        return result;
```
kerrrusha, line 63: we can use T-W-R for reading operation

src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +46,20 @@ public Optional<Movie> get(Long id) {
...
+            final Query<Movie> query = session.createQuery(sqlText, Movie.class);
+            result = query.list();
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't get a list of movies.", e);
+        }
+        return result;
```
kerrrusha, line 63: we can use T-W-R for reading operation

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,70 @@
...
+            query.setParameter("session_date", date);
+            result = query.list();
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't get a list of movie sessions.", e);
+        }
+        return result;
```
kerrrusha, line 68: we can use T-W-R for reading operation

src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,53 @@
...
+@Table(name = "cinema_hall")
+public class CinemaHall {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private int capacity;
+    private String description;
+
+    public CinemaHall() {
+    }
```
kerrrusha, line 19: redundant explicit constructor declaration
```suggestion
```

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,74 @@
...
+    private Movie movie;
+    @ManyToOne()
+    @JoinColumn(name = "cinemahall_id", foreignKey = @ForeignKey(name =
+            "FK_movieSession_to_cinemaHall"))
+    private CinemaHall cinemaHall;
+    @Column(name = "show_time")
+    private LocalDateTime showTime;
+
+    public MovieSession() {
+    }
```
kerrrusha, line 31: redundant explicit constructor declaration
```suggestion
```

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(final CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(final Long id) {
+        return cinemaHallDao.get(id).get();
```
kerrrusha, line 22: using `Optional.get()` without null-check is a bad practice. consider using `.orElseThrow(() -> new SomeException("message"))` instead

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    @Inject private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(final MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(final Long id) {
+        return movieSessionDao.get(id).get();
```
kerrrusha, line 22: using Optional.get() without null-check is a bad practice. consider using .orElseThrow(() -> new SomeException("message")) instead

kerrrusha (changes requested):
almost good

src/main/java/mate/academy/dao/MovieSessionDao.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.List;
+import java.util.Optional;
+import mate.academy.model.MovieSession;
+
+public interface MovieSessionDao {
+    MovieSession add(MovieSession movie);
+
+    Optional<MovieSession> get(Long id);
+
+    List<MovieSession> findAvailableSessions(final Long movieId, final LocalDate date);
```
Sanderfuck, line 13: Why do you need final?

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,71 @@
...
+        this.cinemaHall = cinemaHall;
+    }
+
+    public LocalDateTime getShowTime() {
+        return showTime;
+    }
+
+    public void setShowTime(final LocalDateTime showTime) {
+        this.showTime = showTime;
+    }
```
Sanderfuck, line 60: Why you make parameters final?

Sanderfuck (changes requested):

src/main/java/mate/academy/model/MovieSession.java
Krasnov-Maksim: I can remove the word 'final'.
But I read the articles below and decided that if I do not change the argument inside the method, then it is useful to declare argument final.
Or is this a mistake?

https://stackoverflow.com/questions/500508/why-should-i-use-the-keyword-final-on-a-method-parameter-in-java
Never Reassign Arguments
As good programming practice (in any language), you should never re-assign a parameter/argument variable to an object other than the object passed by the calling method. In the examples above, one should never write the line arg = . Since humans make mistakes, and programmers are human, let’s ask the compiler to assist us. Mark every parameter/argument variable as 'final' so that the compiler may find and flag any such re-assignments.
-------------------------------------------------------------------------------

https://softwareengineering.stackexchange.com/questions/48413/in-java-should-i-use-final-for-parameters-and-locals-even-when-i-dont-have-t#:~:text=Java%20allows%20marking%20variables%20(fields,are%20meant%20to%20be%20immutable.

This is exactly why you should use the final keyword. You state that YOU know it'll never be re-assigned, but no one else knows that. Using final immediately disambiguates your code that tiny bit more.

sokimaaa (approved):

PR #726, opened 2023-11-17
added movie session and cinema hall entities also dao and service lay…

Andriy-Hrynevetskyi:
…ers for them.

Implemented all necessary methods above.

### DIFF
--- a/src/main/java/mate/academy/Main.java
+++ b/src/main/java/mate/academy/Main.java
@@ -1,16 +1,59 @@
 package mate.academy;
 
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.List;
+import mate.academy.lib.Injector;
+import mate.academy.model.CinemaHall;
 import mate.academy.model.Movie;
+import mate.academy.model.MovieSession;
+import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
+import mate.academy.service.MovieSessionService;
 
 public class Main {
     public static void main(String[] args) {
-        MovieService movieService = null;
+        Injector injector = Injector.getInstance("mate.academy");
+        MovieService movieService = (MovieService) injector.getInstance(MovieService.class);
 
         Movie fastAndFurious = new Movie("Fast and Furious");
         fastAndFurious.setDescription("An action film about street racing, heists, and spies.");
         movieService.add(fastAndFurious);
         System.out.println(movieService.get(fastAndFurious.getId()));
         movieService.getAll().forEach(System.out::println);
+
+        CinemaHallService cinemaHallService
+                = (CinemaHallService) injector.getInstance(CinemaHallService.class);
+
+        CinemaHall hall = new CinemaHall();
+        hall.setCapacity(300);
+        hall.setDescription("King size hall)");
+        cinemaHallService.add(hall);
+        System.out.println(cinemaHallService.get(hall.getId()));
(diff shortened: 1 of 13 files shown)


src/main/java/mate/academy/Main.java
```diff
@@ -1,16 +1,59 @@
...
 import mate.academy.model.Movie;
+import mate.academy.model.MovieSession;
+import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
+import mate.academy.service.MovieSessionService;
 
 public class Main {
     public static void main(String[] args) {
-        MovieService movieService = null;
+        Injector injector = Injector.getInstance("mate.academy");
```
bhdnchui, line 16: I think it would be better to make it as constant.

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,64 @@
...
+import jakarta.persistence.Table;
+import java.time.LocalDateTime;
+
+@Entity
+@Table(name = "movie_sessions")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
```
bhdnchui, line 17: Add explicit mapping here.
With `@JoinColumn` annotation.

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,64 @@
...
+    private Movie movie;
+    @ManyToOne
```
bhdnchui, line 19: Same.

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).orElseThrow(
+                () -> new NoSuchElementException("Can`t get the cinema hall")
```
bhdnchui, line 24: Expand your error message with id.

bhdnchui: It's better to inform user that `cinema hall by id not found`. Message with that construction is better for user because it explains user that we can find a cinema hall, but the cinema hall with that id does not exist.

bhdnchui: Make sure to do it everywhere.

src/main/java/mate/academy/dao/MovieSessionDao.java
```diff
@@ -0,0 +1,13 @@
+package mate.academy.dao;
+
+import java.time.LocalDate;
+import java.util.List;
+import mate.academy.model.MovieSession;
+
+public interface MovieSessionDao {
+    MovieSession add(MovieSession movieSession);
+
+    MovieSession get(Long id);
```
bhdnchui, line 10: Return optional here.

bhdnchui (changes requested):
Good job! Just a few points to consider.

kshuryhin (approved):
Looks good to me

PR #746, opened 2023-12-28
HW hibernate-create-movie-session-hw

arrem4eg:


src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -38,13 +38,17 @@ public Movie add(Movie movie) {
...
+        } catch (RuntimeException e) {
             throw new DataProcessingException("Can't get a movie by id: " + id, e);
         }
     }
 
     @Override
     public List<Movie> getAll() {
-        return null;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("from Movie", Movie.class).getResultList();
```
Ivan95kos, line 49:
```suggestion
            return session.createQuery("FROM Movie", Movie.class).getResultList();
```

Ivan95kos (approved):
Great job, with a minor comment

PR #753, opened 2024-01-06
implementations of Movie, MoviesSession and CinemaHall classes and it…

ivanlavrinenko:
…'s services and daos


src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,57 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a cinema hall by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<CinemaHall> getAllCinemaHallQuery = session
+                    .createQuery("from CinemaHall", CinemaHall.class);
```
Ivan95kos, line 51: It is better to write all SQL commands in upper case 
```suggestion
                    .createQuery("FROM CinemaHall", CinemaHall.class);
```

src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +46,11 @@ public Optional<Movie> get(Long id) {
 
     @Override
     public List<Movie> getAll() {
-        return null;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Movie> getAllMoviesQuery = session.createQuery("from Movie", Movie.class);
```
Ivan95kos, line 50: the same

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public MovieSession get(Long id) {
+        Optional<MovieSession> optionalMovieSession = movieSessionDao.get(id);
+        if (optionalMovieSession.isPresent()) {
+            return optionalMovieSession.get();
+        } else {
+            throw new DataProcessingException("Movie session with id "
+                    + id + " does not exist in database");
+        }
```
Ivan95kos, line 31: It is better to use the `orElseThrow` method

Ivan95kos (approved):

PR #763, opened 2024-01-19
solution 19/01

VadimKadasiev:


src/main/java/mate/academy/service/impl/MovieServiceImpl.java
```diff
@@ -19,11 +20,13 @@ public Movie add(Movie movie) {
 
     @Override
     public Movie get(Long id) {
-        return movieDao.get(id).get();
+
+        return movieDao.get(id).orElseThrow(() -> new DataProcessingException(
+                "Can't get movie from DB by ID " + id, new RuntimeException()));
```
okuzan, line 25: we use DataProcessing on DAO level, on Service level let's use EntityNotFound
don't create new Runtime just for the sake of it, if there's nothing to rethrow, that's ok
Fix everywhere

okuzan (changes requested):

src/main/java/mate/academy/dao/CinemaHallDao.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.Optional;
+import mate.academy.model.CinemaHall;
+
+public interface CinemaHallDao {
+    CinemaHall add(CinemaHall cinemaHall);
+
+    Optional<CinemaHall> get(Long id);
+
+    List<CinemaHall> getAll();
+
```
okuzan, line 13:
```suggestion
```

okuzan (approved):

PR #772, opened 2024-02-02
hm for review

vdbond:
main method was tested

### DIFF
--- a/src/main/java/mate/academy/dao/CinemaHallDao.java
+++ b/src/main/java/mate/academy/dao/CinemaHallDao.java
@@ -1,0 +1,12 @@
+package mate.academy.dao;
+
+import java.util.List;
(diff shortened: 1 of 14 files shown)


src/main/java/mate/academy/dao/CinemaHallDao.java
```diff
@@ -0,0 +1,12 @@
+package mate.academy.dao;
+
+import java.util.List;
+import mate.academy.model.CinemaHall;
+
+public interface CinemaHallDao {
+    CinemaHall add(CinemaHall cinemaHall);
+
+    CinemaHall get(Long id);
```
olekskov, line 9: same here

olekskov (changes requested):

kozhukhovsky (approved):

PR #776, opened 2024-02-09
hw create movie session

JaysonJuza:


src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">user</property>
+        <property name="connection.password">password</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.model.Movie"/>
+        <mapping class="mate.academy.model.CinemaHall"/>
+        <mapping class="mate.academy.model.MovieSession"/>
+    </session-factory>
+</hibernate-configuration>
```
okuzan, line 15:
```suggestion
</hibernate-configuration>

```
each file should end with one empty line, which will not be displayed on github, if there is not line, a red symbol like here will appear

src/main/java/mate/academy/dao/MovieSessionDao.java
```diff
@@ -0,0 +1,15 @@
+package mate.academy.dao;
+
+import java.time.LocalDate;
+import java.util.List;
+import java.util.Optional;
+import mate.academy.model.MovieSession;
+
+public interface MovieSessionDao {
+
```
okuzan, line 9: no empty lines after class/method definition

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public Optional<MovieSession> get(Long id) {
+        return movieSessionDao.get(id);
```
okuzan, line 28: using get on optional without checking it first is a bad practice
use orElseThrow + entityNotFound + message + id

okuzan (changes requested):

Rommelua (approved):

PR #785, opened 2024-02-22
Hibernate create movie session hw

AdamMudrak:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).orElseThrow();
```
okuzan, line 23: throw entity not found exception + id in message everywhere

okuzan (changes requested):

okuzan (approved):

PR #786, opened 2024-02-27
done

163onmyneck:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
okuzan, line 23: bad practice, what's the point of using optional if the code can still produce NPE
use orElseThrow + EntityNotFound + message + id everywhere

okuzan (changes requested):

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).orElseThrow(() ->
+                new NoSuchElementException("Cinema hall with this id doesn't exist"));
```
kozhukhovsky, line 24: add id to message

src/main/java/mate/academy/service/impl/MovieServiceImpl.java
```diff
@@ -19,11 +20,12 @@ public Movie add(Movie movie) {
 
     @Override
     public Movie get(Long id) {
-        return movieDao.get(id).get();
+        return movieDao.get(id).orElseThrow(() ->
+                new NoSuchElementException("Movie with this id doesn't exist"));
```
kozhukhovsky, line 24: same

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).orElseThrow(() ->
+                new NoSuchElementException("Movie session with this id doesn't exist"));
```
kozhukhovsky, line 25: same

kozhukhovsky (approved):

PR #787, opened 2024-02-28
Created CinemaHall and MovieSession classes, added implementations to…

misha-levko:
… MovieDaoImpl and MovieServiceImpl, created CinemaHallDaoImpl and CinemaHallServiceImpl, created instances of new services with using injector.

### DIFF
--- a/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
@@ -1,11 +1,12 @@
 package mate.academy.dao.impl;
 
+import java.time.LocalDate;
 import java.util.List;
 import java.util.Optional;
-import mate.academy.dao.MovieDao;
+import mate.academy.dao.MovieSessionDao;
 import mate.academy.exception.DataProcessingException;
 import mate.academy.lib.Dao;
-import mate.academy.model.Movie;
+import mate.academy.model.MovieSession;
 import mate.academy.util.HibernateUtil;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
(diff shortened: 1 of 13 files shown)


src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+    private final CinemaHallDao cinemaHallDao = new CinemaHallDaoImpl();
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
okuzan, line 21: bad practice, what's the point of using optional if the code can still produce NPE
use orElseThrow + EntityNotFound + message + id everywhere

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,59 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a movie by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+        try (Session session = sessionFactory.openSession()) {
+            Query<MovieSession> getMovieSessionQuery = session.createQuery("FROM MovieSession m "
```
okuzan, line 53:
```suggestion
            Query<MovieSession> getMovieSessionQuery = session.createQuery(
                                "FROM MovieSession m "
```
let's try to make it look like a SQL block

okuzan (changes requested):

PR #789, opened 2024-02-28
First Solution. Implemented requested DAO, services, entities and met…

kacper-kukula:
…hods.

### DIFF
--- a/src/main/java/mate/academy/model/CinemaHall.java
+++ b/src/main/java/mate/academy/model/CinemaHall.java
@@ -7,20 +7,20 @@ import jakarta.persistence.Id;
 import jakarta.persistence.Table;
 
 @Entity
-@Table(name = "movies")
-public class Movie {
+@Table(name = "cinema_halls")
+public class CinemaHall {
 
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
-    private String title;
+    private int capacity;
     private String description;
 
-    public Movie() {
+    public CinemaHall() {
     }
 
-    public Movie(String title) {
-        this.title = title;
+    public CinemaHall(int capacity) {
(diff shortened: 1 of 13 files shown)


src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,61 @@
...
+    private String description;
+
+    public CinemaHall() {
+    }
+
+    public CinemaHall(int capacity) {
+        this.capacity = capacity;
+    }
+
+    public Long getId() {
```
nowickimj, line 26: Why not using lombok for getters, setters, toString?

src/main/java/mate/academy/model/Movie.java
```diff
@@ -9,6 +9,7 @@
 @Entity
 @Table(name = "movies")
 public class Movie {
+
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
```
nowickimj, line 15: For each entity, you use the same identifier type.
How about extracting it to `AbstractEntity` class?

nowickimj (changes requested):
2 minor comments

src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@Entity
+@Table(name = "cinema_halls")
+@Getter
+@Setter
+@NoArgsConstructor
+@ToString
```
nowickimj, line 15: in order to include the super class fields as well (id), you must add `@ToString(callSuper=true)`

nowickimj (approved):

PR #803, opened 2024-03-26
fixed

LizaBalshv:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).orElseThrow(() ->
+                new DataProcessingException("Can`t get MovieSession with id" + id));
```
okuzan, line 24: this exception is for DAO layer, on service layer we use EntityNotFoundException in such cases

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">root</property>
+        <property name="connection.password">Kondrateva1_</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.model.Movie"/>
+        <mapping class="mate.academy.model.MovieSession"/>
+        <mapping class="mate.academy.model.CinemaHall"/>
+    </session-factory>
+</hibernate-configuration>
```
okuzan, line 15: don't forget to leave an empty line at the end of the file

okuzan (changes requested):

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+    private final CinemaHallDao cinemaHallDao = new CinemaHallDaoImpl();
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
MAMentorFX, line 21: use orElseThrow instead of get()

MAMentorFX (approved):

PR #816, opened 2024-04-05
build success

MolchanovAlexander:


src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,50 @@
...
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "cinema_halls")
+public class CinemaHall {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private int capacity;
```
karma-o, line 15: if we are using primitive fields for entity - than we should mark them with `@NotNull` constraint, or use wrappers as their types to avoid the null confusion

karma-o (approved):
Left a minor comment

PR #836, opened 2024-05-14
movie-session-resolved

margosalii:


src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public List<MovieSession> findAvailableSessions(Long movieId, LocalDate date) {
+        LocalDateTime daysBeginning = LocalDateTime.of(date, LocalTime.MIN);
+        LocalDateTime daysEnding = LocalDateTime.of(date, LocalTime.MAX);
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<MovieSession> availableSessionsQuery = session.createQuery("from MovieSession ms "
+                    + "left join fetch ms.movie "
+                    + "where ms.movie.id = :id AND ms.localDateTime "
+                    + "BETWEEN :fromDate AND :toDate", MovieSession.class);
```
okuzan, line 58: either upperase all keywords or lowercase all of them

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
okuzan, line 23: this isn't null safe code, which defeats the very purpose of using optional. use orElseThrow + entity not found + message + id instead

okuzan (changes requested):
fix .get() everywhere

PR #843, opened 2024-06-02
resolve create movie session task

YouraSoliar:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
okuzan, line 23: defeats purpose or using optional - you will get NPE. Use orElseThrow instead to fail fast with a clear message, it's a better practice. Use EntityNotFoundException + message + id

okuzan (changes requested):

Elena-Bruyako (changes requested):
Good job!

Elena-Bruyako (approved):

PR #864, opened 2024-07-31
solved homework

Yaromyros:


src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).orElseThrow(() ->
+                new DataProcessingException("There is no movieSession"));
```
okuzan, line 25: For Service layer use EntityNotFoundException, add id to the message
DataProcessing is for a DAO layer

okuzan (approved):

PR #866, opened 2024-08-04
first attempt

jszymanski97:


src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    @Override
+    public Optional<CinemaHall> get(Long id) {
+        CinemaHall cinemaHall;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            session.beginTransaction();
+            cinemaHall = session.get(CinemaHall.class, id);
+            return Optional.ofNullable(cinemaHall);
```
beatarjn, line 43: You could directly return the expression without creating additional variable
`Optional.ofNullable(session.get(CinemaHall.class, id))`

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,58 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Error getting cinema hall id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            session.beginTransaction();
+            return session.createQuery("from CinemaHall", CinemaHall.class).list();
```
beatarjn, line 53: It's better to use return session.createQuery("from CinemaHall", CinemaHall.class).getResultList();
It's more in line with JPA standards and might be considered slightly more modern and readable.
list() is functionally equivalent to getResultList() but is considered more idiomatic in older Hibernate versions.

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,58 @@
...
+            return Optional.ofNullable(cinemaHall);
+        } catch (Exception e) {
+            throw new DataProcessingException("Error getting cinema hall id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            session.beginTransaction();
```
beatarjn, line 52: You do not need to explicitly begin a transaction. For read operations, especially if you are not modifying data, this approach is often sufficient and preferable because it avoids the overhead of transaction management.
However if you are opening transaction, pls remember to commit/close it.

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,72 @@
...
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<MovieSession> get(Long id) {
+        MovieSession movieSession;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            session.beginTransaction();
```
beatarjn, line 44: The same as above, retrun expression directly and no need to begin transaction

beatarjn (approved):
Looks good!

PR #868, opened 2024-08-07
hibernate-create-movie-session-hw-solution

VArabchuk:


src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,14 @@
...
+        <property name="connection.driver_class">com.mysql.cj.jdbc.Driver</property>
+        <property name="connection.username">root</property>
+        <property name="connection.password">11111111</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+        <mapping class="mate.academy.model.Movie"/>
+        <mapping class="mate.academy.model.CinemaHall"/>
+        <mapping class="mate.academy.model.MovieSession"/>
+    </session-factory>
+</hibernate-configuration>
```
okuzan, line 14:
```suggestion
</hibernate-configuration>

```

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).get();
```
okuzan, line 23: using .get on Optional without checking is a bad practice - it can cause NPE which defeats purpose of optional, use orElseThrow + entity not found + message + id

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).get();
```
okuzan, line 22: same

okuzan (changes requested):

okuzan (changes requested):
DataProcessing is for DAO layer, not the service layer, here, EntityNotFoundException is more suitable

okuzan (approved):

PR #1037, opened 2026-04-21
hibernate-session

weronikarzekiecka:


src/main/java/mate/academy/Main.java
```diff
@@ -1,16 +1,47 @@
...
+        movieSessionService.add(movieSession);
+        System.out.println(movieSessionService.get(1L));
+
+        List<MovieSession> availableSessions =
+                movieSessionService.findAvailableSessions(1L, LocalDate.of(2026, 4, 7));
+        System.out.println("Session of movie " + availableSessions);
+
     }
-}
+}
```
mateuszwojtkowiak: Your project is not building with Maven. You need to add an empty line in Main class. This is the error: 
`Error:  /home/runner/work/hibernate-create-movie-session-hw/hibernate-create-movie-session-hw/src/main/java/mate/academy/Main.java:1: File does not end with a newline. [NewlineAtEndOfFile]`

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery(
+                    "SELECT DISTINCT c FROM CinemaHall c", CinemaHall.class).list();
+        } catch (HibernateException e) {
+            throw new RuntimeException("Can't get all cinema halls", e);
```
mateuszwojtkowiak: Please replace the `RuntimeException` with `DataProcessingException` to keep exception handling consistent across all DAO methods.

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,56 @@
...
+            throw new DataProcessingException("Can't get a cinema hall by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery(
+                    "SELECT DISTINCT c FROM CinemaHall c", CinemaHall.class).list();
+        } catch (HibernateException e) {
```
mateuszwojtkowiak: Catching only `HibernateException` is too narrow here. Other exceptions (e.g., IllegalArgumentException, PersistenceException) may also occur. Please catch a generic `Exception` instead, similar to the other DAO methods.

src/main/java/mate/academy/dao/impl/CinemaHallDaoImpl.java
```diff
@@ -0,0 +1,56 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get a cinema hall by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<CinemaHall> getAll() {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery(
+                    "SELECT DISTINCT c FROM CinemaHall c", CinemaHall.class).list();
```
mateuszwojtkowiak: In HQL the `SELECT` clause is optional when querying the root entity.  
This query can be simplified to a more idiomatic form by removing both  
`SELECT` and `DISTINCT`, since there are no joins that could produce duplicates.

Current version:
`SELECT DISTINCT c FROM CinemaHall c`

Suggested version:
`FROM CinemaHall`

src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +46,10 @@ public Optional<Movie> get(Long id) {
 
     @Override
     public List<Movie> getAll() {
-        return null;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("SELECT DISTINCT m FROM Movie m", Movie.class).list();
+        } catch (HibernateException e) {
```
mateuszwojtkowiak: Catching only `HibernateException` is too narrow here. Other exceptions (e.g., IllegalArgumentException, PersistenceException) may also occur.  
Please catch a generic `Exception` instead, similar to the other DAO methods.

src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +46,10 @@ public Optional<Movie> get(Long id) {
 
     @Override
     public List<Movie> getAll() {
-        return null;
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("SELECT DISTINCT m FROM Movie m", Movie.class).list();
```
mateuszwojtkowiak: In HQL the `SELECT` clause is optional when querying the root entity.  
This query can be simplified to a more idiomatic form by removing both  
`SELECT` and `DISTINCT`, since there are no joins that could produce duplicates.

Current version:
`SELECT DISTINCT m FROM Movie m`

Suggested version:
`FROM Movie`

src/main/java/mate/academy/dao/impl/MovieDaoImpl.java
```diff
@@ -45,6 +46,10 @@ public Optional<Movie> get(Long id) {
...
+        } catch (HibernateException e) {
+            throw new RuntimeException("Can't get all Movies", e);
```
mateuszwojtkowiak: Please replace `RuntimeException` with `DataProcessingException` to keep exception handling consistent across all DAO methods.

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,64 @@
...
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery(
+                            "FROM MovieSession m WHERE m.movie.id = :movieId AND"
+                                    + " m.showTime >= :start AND m.showTime < :nextStart",
+                            MovieSession.class)
+                    .setParameter("movieId", movieId)
+                    .setParameter("start", date.atStartOfDay())
+                    .setParameter("nextStart", date.plusDays(1).atStartOfDay())
+                    .list();
+        } catch (HibernateException e) {
```
mateuszwojtkowiak: Catching only `HibernateException` is too narrow here.  
Other exceptions (e.g., IllegalArgumentException, PersistenceException) may also occur.  
Please catch a generic `Exception` instead, as in the other DAO methods.

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -0,0 +1,64 @@
...
+            throw new RuntimeException(
```
mateuszwojtkowiak: Please replace `RuntimeException` with `DataProcessingException` to keep exception handling consistent across all DAO methods.

src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -0,0 +1,50 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private int capacity;
+    private String description;
+
+    public Long getId() {
+        return id;
+    }
+
+    public void setId(Long id) {
```
mateuszwojtkowiak: The `setId()` method is not needed because the ID is generated by Hibernate.  
Removing it helps prevent accidental modification of the primary key.

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,64 @@
...
+    private Movie movie;
+    @ManyToOne
+    private CinemaHall cinemaHall;
+    private LocalDateTime showTime;
+
+    public Long getId() {
+        return id;
+    }
+
+    public void setId(Long id) {
```
mateuszwojtkowiak: The `setId()` method is not needed because the ID is generated by Hibernate.  
Removing it helps prevent accidental modification of the primary key.

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,64 @@
...
+import jakarta.persistence.Table;
+import java.time.LocalDateTime;
+
+@Entity
+@Table(name = "movie_session")
+public class MovieSession {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
```
mateuszwojtkowiak: It is recommended to explicitly define `@JoinColumn` for `@ManyToOne` relations.  
Hibernate will generate default column names, but adding `@JoinColumn(name = "...")`  
makes the schema more explicit and avoids surprises when naming conventions change.

src/main/java/mate/academy/model/MovieSession.java
```diff
@@ -0,0 +1,64 @@
...
+    public LocalDateTime getShowTime() {
+        return showTime;
+    }
+
+    public void setShowTime(LocalDateTime showTime) {
+        this.showTime = showTime;
+    }
+
+    @Override
+    public String toString() {
```
mateuszwojtkowiak, line 60: Even though `@ManyToOne` is EAGER by default and will not cause LazyInitializationException  
in the current configuration, including associated entities in `toString()` can still be risky.  
If the fetch type is ever changed to LAZY (which is common for performance reasons),  
this method may start throwing LazyInitializationException.  
It is safer to include only the IDs of related entities.

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private CinemaHallDao cinemaHallDao;
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).orElseThrow();
```
mateuszwojtkowiak: Using `orElseThrow()` without a custom exception message makes debugging harder.  
Consider providing a descriptive message or throwing a custom exception  
to make it clear why the lookup failed.

src/main/java/mate/academy/service/impl/MovieServiceImpl.java
```diff
@@ -19,11 +19,11 @@ public Movie add(Movie movie) {
 
     @Override
     public Movie get(Long id) {
-        return movieDao.get(id).get();
+        return movieDao.get(id).orElseThrow();
```
mateuszwojtkowiak: Using `orElseThrow()` without a custom exception message makes debugging harder.  
Consider providing a descriptive message or throwing a custom exception  
to make it clear why the lookup failed.

src/main/java/mate/academy/service/impl/MovieSessionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    private MovieSessionDao movieSessionDao;
+
+    @Override
+    public MovieSession add(MovieSession movieSession) {
+        return movieSessionDao.add(movieSession);
+    }
+
+    @Override
+    public MovieSession get(Long id) {
+        return movieSessionDao.get(id).orElseThrow();
```
mateuszwojtkowiak: Using `orElseThrow()` without a custom exception message makes debugging harder.  
Consider providing a descriptive message or throwing a custom exception  
to make it clear why the lookup failed.

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,21 @@
...
+        "-//Hibernate/Hibernate Configuration DTD 3.0//EN"
+        "http://hibernate.org/dtd/hibernate-configuration-3.0.dtd">
+
+<hibernate-configuration>
+    <session-factory>
+        <property name="hibernate.dialect">org.hibernate.dialect.MySQL8Dialect</property>
+        <property name="connection.url">jdbc:mysql://localhost/cinema?serverTimezone=UTC</property>
+        <property name="connection.driver_class">com.mysql.cj.jdbc.Driver</property>
+        <property name="connection.username">root</property>
+        <property name="connection.password">741252369Wr?</property>
```
mateuszwojtkowiak, line 12: The database password should not be stored directly in `hibernate.cfg.xml`.  
This file is committed to the repository, so exposing credentials here is a security risk.  
Please move the password to a local, ignored configuration file or use environment variables,  
and reference it in the Hibernate config instead of hard‑coding it.

mateuszwojtkowiak (changes requested):
Good job! Please make sure your project builds correctly with Maven.  
Right now you need to add a newline at the end of the Main class to fix the build issue.  
Other than that, please review my comments and apply the required changes.

src/main/java/mate/academy/Main.java
```diff
@@ -1,16 +1,47 @@
...
-        movieService.getAll().forEach(System.out::println);
+
+        CinemaHall oldHall = new CinemaHall();
+        oldHall.setCapacity(100);
+        oldHall.setDescription("Old central hall");
+
+        CinemaHallService cinemaHallService =
+                (CinemaHallService) injector.getInstance(CinemaHallService.class);
+        cinemaHallService.add(oldHall);
+        CinemaHall cinemaHallFromDb = cinemaHallService.get(1L);
```
slade13, line 29: Use saved entity IDs instead of hardcoded IDs in `Main`

In `Main` you use hardcoded IDs:

```
CinemaHall cinemaHallFromDb = cinemaHallService.get(1L);
...
System.out.println(movieSessionService.get(1L));
...
movieSessionService.findAvailableSessions(1L, LocalDate.of(2026, 4, 7));
```

This can break if the database already contains data or if IDs do not start from `1`.

Better:

```
CinemaHall cinemaHallFromDb = cinemaHallService.get(oldHall.getId());
...
System.out.println(movieSessionService.get(movieSession.getId()));
...
movieSessionService.findAvailableSessions(
        fastAndFurious.getId(),
        LocalDate.of(2026, 4, 7)
);
```

This makes the test logic more stable.

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,21 @@
...
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
```
slade13, line 14: Avoid `create-drop` unless it is intentional for local testing:

```
<property name="hbm2ddl.auto">create-drop</property>
```

This means Hibernate will create the schema on startup and drop it when the application shuts down. For learning/testing it may work, but it also means all data will be lost after each run.

For this task, `update` is usually more convenient:



or, in a more production-like setup, schema management should be handled separately.
```suggestion
        <property name="hbm2ddl.auto">update</property>
```

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public CinemaHall add(CinemaHall cinemaHall) {
+        return cinemaHallDao.add(cinemaHall);
+    }
+
+    @Override
+    public CinemaHall get(Long id) {
+        return cinemaHallDao.get(id).orElseThrow(
+                () -> new RuntimeException(
```
slade13, line 24: In `MovieServiceImpl` and `MovieSessionServiceImpl` you use `NoSuchElementException`, but in `CinemaHallServiceImpl` you use plain `RuntimeException`:

```
throw new RuntimeException("Can't get cinema hall by id: " + id)
```

Better to keep it consistent:

```
throw new NoSuchElementException("Can't find cinema hall by id: " + id)
```

slade13 (approved):
Good job, please review my comments.

